Issue #10, opened 2016-07-18
_cjose_jws_verify_sig_rs256 always returns true


zandbelt:
what's up with this line:
https://github.com/cisco/cjose/blob/master/src/jws.c#L827

````
static bool _cjose_jws_verify_sig_rs256(
            cjose_jws_t *jws, 
            const cjose_jwk_t *jwk, 
            cjose_err *err)
{
    return true;
...
````

linuxwolf:
That's clearly a bug.  Possibly from a bad merge, although I can't recall any of those.

zandbelt:
The code that is triggered now doesn't work conform spec, see #12

linuxwolf:
Re-opening until #12 merges.

linuxwolf:
Closed with merge of #12.

PR #12, opened 2016-07-18
fix RS256 signing/verification and add RS384/RS512 and PS384/PS512

zandbelt:
The RS256 code wasn't working with externally provided signed JWTs, the test code would only validate against JWTs that it created itself (with its own wrong signature). I'm not sure why but I've changed the code to use the more regular RSA_sign/RSA_verify functions.

I have included the JWTs from the JWS spec to test against in the test suite so tests against externally created JWTs are now covered.


linuxwolf:
Having some build issues locally, due to timeouts in `test_cjose_jws_self_sign_self_verify_many()`.  Adding a `tcase_set_timeout(tc_jws, 15.0);` seems to take care of it.

linuxwolf:
Yes, travis-ci.org builds needs to be enabled.  Will work on that.

linuxwolf:
Merged as part of bd52d21f18acb1812d1c125ba744590ccf858da0.

Fixes #10.

PR #19, opened 2016-07-22
support importing unsecure JWS (alg=none)

zandbelt:
import only, verify or sign not allowed; alternatively the signature of the `_cjose_jws_validate_hdr` could have an extra bool `allow_none`

### DIFF
--- a/include/cjose/header.h
+++ b/include/cjose/header.h
@@ -37,6 +37,9 @@ extern const char *CJOSE_HDR_CTY;
 /** The Jose "kid" header attribute. */
 extern const char *CJOSE_HDR_KID;
 
+/** The JWA algorithm attribute value for none. */
+extern const char *CJOSE_HDR_ALG_NONE;
+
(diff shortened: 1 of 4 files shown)


linuxwolf:
I'll admit I'm very hesitant to take on support of "alg":"none".  However, since sign and verify don't work, I think that's ok.

Need to complete a review to be sure.

zandbelt:
I understand; I know it has been a vulnerability in at least one implementation that I know of, but there the problem was exactly that the verify function would return successfully.

Unsecured JWTS are part of the JOSE specs - see the spec example that I've included in the tests - and I need it because it is allowed in backchannel OpenID Connect flows and I wish to avoid awkward exception handling.

Both sign and verify would return false for unsecured JWTs,  the only thing that succeeds is import; it would be an application failure to assume anything related to integrity/encryption only after a succesful import, that goes for "none" as well as the other existing algorithms.

PR #22, opened 2016-07-28
various build fixes

zandbelt:
to allow building/packaging on various platforms


linuxwolf:
Do you have Dockerfiles or images you can share for my edification?  Privately is preferable with me, unless you're ready to share globally.

PR #25, opened 2016-07-28
add support for OpenSSL 1.1.x

zandbelt:
that's all I have/need...!

### DIFF
--- a/include/cjose/util.h
+++ b/include/cjose/util.h
@@ -58,8 +58,11 @@ typedef void (* cjose_dealloc3_fn_t)(void *, const char *, int);
  * is used.
  *
  * \param alloc [in] The custom allocator function to use.
+ * \param alloc3 [in] The custom allocator function to use for OpenSSL >= 1.1.0, called with extra file/line params.
  * \param realloc [in] The custom reallocator function to use.
+ * \param realloc3 [in] The custom reallocator function to use for OpenSSL >= 1.1.0, called with extra file/line params.
  * \param dealloc [in] The custom deallocator function to use.
+ * \param dealloc3 [in] The custom deallocator function to use for OpenSSL >= 1.1.0, called with extra file/line params.
  */
 void cjose_set_alloc_funcs(cjose_alloc_fn_t alloc,
                            cjose_alloc3_fn_t alloc3,
--- a/include/cjose/util.h
+++ b/include/cjose/util.h
@@ -50,7 +50,7 @@ typedef void (* cjose_dealloc3_fn_t)(void *, const char *, int);
 /**
  * Sets the allocator and deallocator functions.
  *
- * If <tt>alloc</tt> is NULL, any previously set allocator function is clared
+ * If <tt>alloc</tt> is NULL, any previously set allocator function is cleared
  * and the the default allocator <tt>malloc()</tt>
  * is used.
  *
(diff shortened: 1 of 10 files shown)


src/jwe.c
```diff
@@ -735,7 +741,12 @@ static bool _cjose_jwe_encrypt_dat_a256gcm(
     }
 
     // get the GCM-mode authentication tag
-    if (EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_GET_TAG, 
+    if (EVP_CIPHER_CTX_ctrl(ctx,
+#if (CJOSE_OPENSSL_11X)
```
linuxwolf: this would be better managed by using a local variable that's set based on the CJOSE_OPENSSL_11X define, instead of #ifdef-ing within the function call

src/jws.c
```diff
@@ -313,13 +313,21 @@ static bool _cjose_jws_build_dig_hmac_sha(
...
     // instantiate and initialize a new mac digest context
+#if (CJOSE_OPENSSL_11X)
+    ctx = HMAC_CTX_new();
+#else
     ctx = cjose_get_alloc()(sizeof(HMAC_CTX));
+#endif
     if (NULL == ctx)
     {
-        CJOSE_ERROR(err, CJOSE_ERR_CRYPTO);
+        CJOSE_ERROR(err, CJOSE_ERR_NO_MEMORY);
```
linuxwolf: good catch!

src/jws.c
```diff
@@ -313,13 +313,21 @@ static bool _cjose_jws_build_dig_hmac_sha(
...
         goto _cjose_jws_build_dig_hmac_sha_cleanup;
     }
+
+#if (CJOSE_OPENSSL_11X)
```
linuxwolf: I think there should be a comment here about why an empty `#if` is really ok.  I can imagine this inadvertently being removed in the future without some indication of why it's there.

include/cjose/util.h
```diff
@@ -53,8 +62,11 @@ typedef void (* cjose_dealloc_fn_t)(void *);
  * \param dealloc [in] The custom deallocator function to use.
```
linuxwolf: missing param comments for (re|de|)alloc3

linuxwolf:
*[@zandbelt looking for discussion before actual code changes]*

The more I go through the changes to `cjose_set_alloc_funcs()`, the less comfortable I am.

I wonder if it would be acceptable to separate out the "simple" alloc|realloc|dealloc functions from the file+line alloc|realloc|dealloc functions.  For example:

```
void cjose_set_alloc_funcs(cjose_alloc_fn_t, cjose_realloc_fn_t, cjose_dealloc_fn_t);
void cjose_set_alloc_ex_funcs(cjose_alloc3_fn_t, cjose_realloc3_fnt_t, cjose_dealloc3_fn_t);
```

Further, I'd be more inclined to define default implementations of cjose_(de|re|)alloc3_fn_t that can be passed to OpenSSL 1.1.x's `CRYPTO_set_mem_functions()`, even if this means real file/line values are lost (unless you call the _ex version).

To me, that eliminates the breaking change to `cjose_set_alloc_funcs()` while still supporting OpenSSL 1.1.x's allocators.

zandbelt:
yes, that makes sense; I did not like the changes much myself either; stay tuned

linuxwolf:
I'm getting compiler errors, but have a branch with the fixes: https://github.com/linuxwolf/cjose/tree/openssl-1.1.x

But I am a little concerned about the dev experience of `cjose_set_alloc_ex_funcs()`.  It feels really clunky, and I wonder if developers will get really turned off with this model.  It seems to me that it would make for a better experience to require the user to either set "basic" allocator/deallocator sets **OR** set "enhanced" allocator/deallocator sets, rather than both.

I'm going to put together a throwaway prototype to express what I mean.

linuxwolf:
Here is the changeset for a mutually exclusive `cjose_get_alloc_funcs()`/`cjose_get_alloc_ex_funcs()`: https://github.com/linuxwolf/cjose/commit/8b82d3085ee3e00c95eeb2bb4432472fc4290783

Let me know what you think @zandbelt

zandbelt:
that looks better indeed

PR #54, opened 2017-06-22
Add NULL terminator to the end of the buffer.

srodman7689:
I added a NULL terminator to the end of the buffer so that other functions expecting the NULL terminator would see it. Not having the NULL terminator was causing issues with JSON deserialization for some JWEs we where working with.

### DIFF
--- a/src/base64.c
+++ b/src/base64.c
@@ -139,6 +139,8 @@ static inline bool _decode(const char *input, size_t inlen, uint8_t **output, si
         buffer[pos++] = (packed >> 16) & 0xff;
     }
 
+    buffer[pos] = '\0';
+
     *output = buffer;
     *outlen = pos;
     assert(*outlen <= rlen);


veselov:
If there are JWEs that don't work, can you please add them to the test suite, so that we can both understand what is the negative outcome, and there is a regression test?

srodman7689:
After going back through what we where doing, we found out that there wasn't really a bug in the code because the lengths of strings are used everywhere. However, we thought we saw a bug because when printing out the raw header inside the cjose_jwe_import function while trying to figure out what was wrong with our code we saw that if you just printf the string then you see a lot of extra random characters on the end of the json because the string is not null terminated. So, if for some reason the strings that are created by decode are used with a function that expects null termination then this will be a problem, but it is not current a problem.

linuxwolf:
Given that the problem is in your application code and this patch in itself makes assumptions that are not always valid (the output is not always a character string), I'm closing this.

PR #55, opened 2017-06-28
Support multiple recipients and JSON serialization for JWE

veselov:
Here is first stab at it. Please take a look and let me know what should be changed, or if there any questions and/or concerns. The formatting changes are due to clang-format (3.9.0 was used).


include/cjose/header.h
```diff
@@ -137,12 +137,43 @@ bool cjose_header_set(cjose_header_t *header, const char *attr, const char *valu
...
+ * \param header[in] a header object.
+ * \param attr[in] the header attribute to be got.
+ * \param err [out] An optional error object which can be used to get additional
+ *        information in the event of an error.
+ * \returns a string containing the current JSON value for the requested attribute.
+ * The value returned is a null terminated UTF-8 encoded JSON string, or NULL if corresponding
+ * header was not found. The returned value is allocated when this function is called, and must
+ * be freed by the caller.
+ */
+char *cjose_header_get_raw(cjose_header_t *header, const char *attr, cjose_err *err);
```
linuxwolf: this should return `const char *` not `char *`.

include/cjose/jwe.h
```diff
@@ -42,21 +44,45 @@ typedef struct _cjose_jwe_int cjose_jwe_t;
...
+ *        is retained by JWE and should be released by the caller if no longer needed.
+ * \param unprotected_header [in] additional header values to include in the shared JWE unprotected header,
+ *        can be NULL. The header is retained by JWE and should be released by the caller if no longer needed.
+ * \param plaintext [in] the plaintext to be encrypted in the JWE payload.
+ * \param plaintext_len [in] the length of the plaintext.
+ * \param err [out] An optional error object which can be used to get additional
+ *        information in the event of an error.
+ * \returns a newly generated JWE with the given plaintext as the payload.
+ */
+cjose_jwe_t *cjose_jwe_encrypt_full(const cjose_jwk_t **jwk,
```
linuxwolf: I think I'd prefer to call this `cjose_jwe_encrypt_multi`, even thought it can take operate with a single recipient.

linuxwolf: Instead of two separate arrays that have to match up, what about making a public `cjose_jwe_recipient_t` struct that has the `jwk` and `unprotected_header`?

include/cjose/jwe.h
```diff
@@ -82,21 +121,55 @@ char *cjose_jwe_export(cjose_jwe_t *jwe, cjose_err *err);
...
+ * \param jwk [in] key_locator callback for finding keys
+ * \param data [in] custom data argument that is passed to the key locator callback.
+ * \param content_len [out] The number of bytes in the returned buffer.
+ * \param err [out] An optional error object which can be used to get additional
+ *        information in the event of an error.
+ * \returns The decrypted content. Note the caller is responsible for free'ing
+ *        this buffer when no longer in use.  Failure to do so will result in
+ *        a memory leak.
+ */
+uint8_t *cjose_jwe_decrypt_full(cjose_jwe_t *jwe, cjose_key_locator key_locator, void *data, size_t *content_len, cjose_err *err);
```
linuxwolf: as with `_encrypt` above, I'd prefer the name be `cjose_jwe_decrypt_multi`.

linuxwolf:
I am very sorry for not getting to this sooner.  Thanks very much for the effort!

include/cjose/header.h
veselov: For this one, the user has to free the result, the value is not handled by the library.

veselov:
@linuxwolf Updated

linuxwolf:
Thanks again for the contribution @veselov !  Sorry for the delays ... work changes and sickness left some of us more behind than we'd like.

PR #58, opened 2017-10-05
LibreSSL support.

tsahara:
With this small change, cjose works well with [LibreSSL](https://www.libressl.org).  All tests pass on OpenBSD 6.1 with bundled LibreSSL and macOS 10.13 with LibreSSL 2.5.5 installed via brew.


linuxwolf:
LGTM!  Thanks!

PR #67, opened 2018-03-05
Prevent segmentation fault when using A256KW + A256GCM

### LINKED ISSUE #66: Bad JWE crashes _cjose_jwe_set_cek_a256gcm
Attempts to decrypt the JWE:
```
eyJhbGciOiJBMjU2S1ciLCJraWQiOiIzYmJhNTkxMC04MzZmLTRjOTYtYjE0My0xNWM0MTkzNTRjNDAiLCJlbmMiOiJBMjU2R0NNIn0.-rqxPUoKNyVTBumul_8dLhK2uVhPRd-l55Xw6No8Fo8wN9wreuQg_g.2ZH9Jxx8tEZ7Dzdz.v2ESxQ.wKR8Ms6EI_st6Pzz424O-w
```
Using JWK:
```
{
    "kty" : "oct",
    "k" : "7SOtIzBXT6MJHRG79G7r3FiIOlNa_tuJX00IzLeYRYY"
}
```
Results in a crash here:
https://github.com/cisco/cjose/blob/master/src/jwe.c#L385

The combined use of alg=A256KW and enc=A256GCM appears to lead to the dereferencing of an invalid pointer (the jwk pointer above has value "1")

linuxwolf:
Fixes #66 

There was an abuse of a pointer in one place that tried to treat it as a boolean flag; the abuse had unintended side effects when certain algorithm choices were used in concert.

There are other inconsistencies within some of the AES-CBC flows, but this takes care of the most egregious.

### DIFF
--- a/src/jwe.c
+++ b/src/jwe.c
@@ -25,10 +25,12 @@
 #include "include/jwe_int.h"
 #include "include/util_int.h"
 
-////////////////////////////////////////////////////////////////////////////////
-static bool _cjose_jwe_set_cek_a256gcm(cjose_jwe_t *jwe, const cjose_jwk_t *jwk, cjose_err *err);
 
-static bool _cjose_jwe_set_cek_aes_cbc(cjose_jwe_t *jwe, const cjose_jwk_t *jwk, cjose_err *err);
+
+////////////////////////////////////////////////////////////////////////////////
+static bool _cjose_jwe_set_cek_a256gcm(cjose_jwe_t *jwe, const cjose_jwk_t *jwk, bool prealloc, cjose_err *err);
+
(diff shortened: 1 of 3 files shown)


src/jwe.c
```diff
@@ -538,7 +540,7 @@ _cjose_jwe_decrypt_ek_aes_kw(_jwe_int_recipient_t *recipient, cjose_jwe_t *jwe,
     }
 
     // generate empty CEK so the the right amount of memory is allocated (abuse JWK parameter to empty)
```
balthorium: Comment about abusing JWK parameter probably no longer needed.

PR #75, opened 2018-06-21
Fix compilation error on Windows in src/concatkdf.c

tgorol:


src/concatkdf.c
```diff
@@ -7,12 +7,25 @@
...
 #include <arpa/inet.h>
+#include <alloca.h>
+#endif
 #include <openssl/evp.h>
 #include <string.h>
 #include <cjose/base64.h>
 #include <cjose/util.h>
 
+#ifdef _WIN32
+#define StackAlloc _alloca
```
linuxwolf: I'd prefer to keep with the convention of single case; for macros that has been all upper case.

PR #80, opened 2018-09-17
Cleanup some warnings about \param lines in header files

jogu:
Xcode warns about these as they describe parameters that don't exist;
add spaces (as done in other \param lines) between the parameter name
and the [in] / [out].

Also correct one parameter name that no longer matches the actual name.

### DIFF
--- a/include/cjose/header.h
+++ b/include/cjose/header.h
@@ -110,7 +110,7 @@ cjose_header_t *cjose_header_new(cjose_err *err);
  * header will be used past the scope it was created in (e.g., from a
  * `cjose_jws_t` object).
  *
- * \param header[in] the header object to be retained.
+ * \param header [in] the header object to be retained.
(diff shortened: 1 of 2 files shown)


zandbelt:
Assessment while merging the OpenIDC/cjose fork back (see #142): unlike most of the older open PRs, this one is not yet on `master`. `include/cjose/header.h` still has twelve `\param header[in]`-style lines without the space, and `include/cjose/jwk.h` still documents the RSA key-size parameter as `size` while the function calls it `keysize`. The OpenIDC fork has carried this change since 2018 (OpenIDC/cjose#1, thanks @jogu).

The diff still applies cleanly to current `master` (`git apply --check` passes). I deliberately left these hunks out of #152 to avoid duplicating an open PR, so it would be good to merge this one as it stands.

TheStormN (approved):

PR #82, opened 2018-11-09
Minor updates for conformance

ajishna:
Minor updates for standards conformance issues (const qualifiers differ in .c and .h files) found when building with VS2017.
jws.c files are updated to use the cjose dealloc function instead of free so it points to the right function pointer when allocating memory functions through cjose_set_alloc_funcs()


TheStormN:
Fixed by #96

PR #87, opened 2019-04-25
Support FIPS-mode in cjose_jwk_derive_ecdh_ephemeral_key

deeringc:
Adding an optional salt parameter to cjose_jwk_derive_ecdh_ephemeral_key

In OpenSSL FIPS-mode, calling the HMAC API with an empty salt results in an error.
In order to maintain current behaviour, pass in an empty string as a salt.
Also missing in this code was checking the return value from the call to HMAC.

### DIFF
--- a/include/cjose/jwk.h
+++ b/include/cjose/jwk.h
@@ -357,17 +357,27 @@ cjose_jwk_t *cjose_jwk_import_json(cjose_header_t *json, cjose_err *err);
  *
  * \param jwk_self [in] The caller's own EC key pair.
  * \param jwk_peer [in] The peer's EC public key.
+ * \param salt [in] An optional salt to apply to the HMAC calculation. Unless FIPS mode is required this can be empty.
+ * \param salt_len [in] The length of the optional salt.
  * \param err [out] An optional error object which can be used to get additional
  *        information in the event of an error.
  * \returns A new JWK representing the ephemeral key, or NULL in the event of
  *        and error.
  */
-cjose_jwk_t *cjose_jwk_derive_ecdh_ephemeral_key(const cjose_jwk_t *jwk_self, const cjose_jwk_t *jwk_peer, cjose_err *err);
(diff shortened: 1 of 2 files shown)


include/cjose/jwk.h
```diff
@@ -26,7 +26,8 @@ extern "C" {
 #endif
 
 /** Enumeration of supported JSON Web Key (JWK) types */
-typedef enum {
+typedef enum
```
deeringc: clang-format did this

src/jwk.c
```diff
@@ -1865,11 +1863,19 @@ bool cjose_jwk_hkdf(const EVP_MD *md,
     // HKDF-Extract, HMAC-SHA256(salt, IKM) -> PRK
     unsigned int prk_len;
     unsigned char prk[EVP_MAX_MD_SIZE];
-    HMAC(md, salt, salt_len, ikm, ikm_len, prk, &prk_len);
+    if (NULL == HMAC(md, salt, salt_len, ikm, ikm_len, prk, &prk_len))
```
deeringc: This missing error handling led to some confusion, because the overall cjose_err reported no error while in FIPS mode but the derivation had actually failed.

deeringc: https://www.openssl.org/docs/man1.1.0/man3/HMAC.html

> HMAC() returns a pointer to the message authentication code or NULL if an error occurred.

src/jwk.c
```diff
@@ -1865,11 +1863,19 @@ bool cjose_jwk_hkdf(const EVP_MD *md,
...
+        CJOSE_ERROR(err, CJOSE_ERR_CRYPTO);
+        return false;
+    }
 
     // HKDF-Expand, HMAC-SHA256(PRK,0x01) -> OKM
     const unsigned char t[] = { 0x01 };
-    HMAC(md, prk, prk_len, t, sizeof(t), okm, NULL);
+    if (NULL == HMAC(md, prk, prk_len, t, sizeof(t), okm, NULL))
+    {
+        CJOSE_ERROR(err, CJOSE_ERR_CRYPTO);
```
deeringc: Is this the right error code?

PR #89, opened 2019-05-08
Fixes #88

### LINKED ISSUE #88: Decoding multi without matching key causes a SEGV
Calling `cjose_jwe_decrypt_multi`, and not having any keys match, causes a SEGV, as an attempt to decrypt with `null` CEK is attempted

veselov:
After key matching is attempted, check that JWE object has any CEK at all, return error if it doesn't.


TheStormN:
Closing if favor of #146

Issue #98, opened 2019-06-26
clang-format status


rnapier:
CONTRIBUTING indicates that before submitting, clang-format must be run (specifically 3.9), but many of the files do not appear to be consistent with clang-format currently. Is clang-format still required and does it require specifically 3.9 (the current version is 8)? I believe this will generate a large number of unrelated changes.

PR #99, opened 2019-06-26
Encode JWS messages with compact JSON

rnapier:
Currently, the JOSE header is encoded with unnecessary spaces. This makes the compact serialization larger than it needs to be. This requests compact JSON from Jansson.

Note that I have not run clang-format on this (see #98), but have run tests.


rnapier:
(I apologize, this PR included other PRs which wasn't intended.)

PR #102, opened 2019-06-26
Extract cjose_jwe_encrypt_iv to allow explicit IV

rnapier:
It is currently not possible to pass a static IV to the encryption functions. This breaks systems that use a counter for GCM mode (rather than a random nonce). This PR extracts encryption functions that permit passing an explicit IV. It is backward compatible with the existing functions, and if NULL is passed, the existing IV-generation functions will be used.


TheStormN:
Closing in favor of #153

PR #104, opened 2019-07-23
Fix buffer overflow in test_cjose_jwe_multiple_recipients().

mpsun:
The function allocates two cjose_jwe_recipient_t on the stack and then writes at index 2, which is out of bounds. The NULL pointer also seems unnecessary because cjose_jwe_encrypt_multi() takes a count as its argument.


TheStormN:
Fixed by #140

PR #117, opened 2021-08-10
jws: _cjose_jws_build_dig_hmac_sha: fix jws->dig reuse memory leak fix

traeak:
This fixes a memory leak already addressed in _cjose_jws_build_dig_sha when a jws is reused for a validation.


TheStormN:
Closing in flavor of #147

PR #126, opened 2023-07-12
use fixed authentication tag length of 16 octets in AES GCM decryption

### LINKED ISSUE #125: SECURITY VULNERABILITY: incorrect Authentication Tag length usage in AES GCM decryption
**NOTE THAT AES GCM DECRYPTION IS SEVERELY BROKEN FOR ALL VERSIONS OF CJOSE < 0.6.2.2**

The AES GCM decryption routine incorrectly uses the Tag length from the actual Authentication Tag provided in the JWE, see: https://github.com/cisco/cjose/blob/0.6.1/src/jwe.c#L1228-L1229:
```
  // set the expected GCM-mode authentication tag
  if (EVP_CIPHER_CTX_ctrl(ctx, CJOSE_EVP_CTRL_GCM_SET_TAG, jwe->enc_auth_tag.raw_len, jwe->enc_auth_tag.raw) != 1)
```
However, the spec https://datatracker.ietf.org/doc/html/rfc7518#section-4.7 says that a fixed length of 16 octets must be applied:
> The requested size of the Authentication Tag output MUST be 128 bits, regardless of the key size.

Therefore this bug allows an attacker to provide a truncated Authentication Tag and to modify the JWE accordingly.

See also: CVE-2023-37464 and https://github.com/OpenIDC/cjose/security/advisories/GHSA-3rhg-3gf2-6xgj

zandbelt:
closes: https://github.com/cisco/cjose/issues/125

### DIFF
--- a/src/jwe.c
+++ b/src/jwe.c
@@ -1225,6 +1225,12 @@ static bool _cjose_jwe_decrypt_dat_a256gcm(cjose_jwe_t *jwe, cjose_err *err)
         goto _cjose_jwe_decrypt_dat_a256gcm_fail;
     }
 
+    if (jwe->enc_auth_tag.raw_len != 16)
+    {
+        CJOSE_ERROR(err, CJOSE_ERR_CRYPTO);
+        goto _cjose_jwe_decrypt_dat_aes_gcm_fail;
+    }
+
     // set the expected GCM-mode authentication tag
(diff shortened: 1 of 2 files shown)


TheStormN:
Fixed by this cherry-pick PR #134 from your fork.

PR #131, opened 2026-04-27
Bugfix/hardening fixes

bifurcation:
This PR implements a number of hardening fixes suggested by Claude Code.  It also converts some declarations from `foo()` to `foo(void)` so that the build passes on platforms with `-Wstrict-prototypes` enabled by default.

I have verified that this branch builds and passes tests locally.

### DIFF
--- a/src/header.c
+++ b/src/header.c
@@ -6,6 +6,7 @@
  */
 
 #include <stdlib.h>
+#include <string.h>
 #include <jansson.h>
 #include "cjose/header.h"
 #include "include/header_int.h"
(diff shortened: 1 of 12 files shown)


src/header.c
```diff
@@ -47,6 +48,59 @@ const char *CJOSE_HDR_EPK = "epk";
 const char *CJOSE_HDR_APU = "apu";
 const char *CJOSE_HDR_APV = "apv";
 
+static const char *CJOSE_HDR_CRIT = "crit";
+
+////////////////////////////////////////////////////////////////////////////////
+bool _cjose_header_validate_crit(cjose_header_t *header, const char *const *supported, size_t supported_len, cjose_err *err)
```
bifurcation, line 54: This validation should be stricter.  RFC 7515 says:

> Producers MUST NOT include Header Parameter names defined by this specification or [[JWA](https://datatracker.ietf.org/doc/html/rfc7515#ref-JWA)] for use with JWS, duplicate names, or names that do not occur as Header Parameter names within the JOSE Header in the "crit" list.  Producers MUST NOT use the empty list "[]" as the "crit" value.

For purposes of this library, that basically means that any header that contains `crit` is invalid.  Since (a) we don't support any non-standard header parameteres, and (b) empty `crit` is forbidden.

balthorium (approved):

PR #140, opened 2026-08-28
Fix Windows build

TheStormN:


test/check_jwe.c
```diff
@@ -1286,8 +1286,6 @@ START_TEST(test_cjose_jwe_multiple_recipients)
         rec[i].unprotected_header = unprotected;
     }
 
-    rec[2].jwk = NULL;
-
```
TheStormN, line 1290: This is referring an non-existing index and thus crashing on Windows. It magically passes on other platforms. I think it is a leftover from a merge conflict, because when I changed the index to be a valid one, the tests were not passing.

As it was never affecting an existing index anyway, removing it seems like the best option here.

CMakeLists.txt
```diff
@@ -46,6 +46,10 @@ endif()
 find_package(OpenSSL 1.0.1 REQUIRED)
 find_package(Jansson 2.3 REQUIRED)
 
+if(WIN32)
+    set(CJOSE_PLATFORM_LIBRARIES ws2_32)
+endif()
```
TheStormN, line 51: I have an idea how to consolidate that to be on a single place, but will do it in a follow up.

CMakeLists.txt
```diff
@@ -113,7 +117,7 @@ if(CJOSE_HAVE_OPENSSL_FEC_H)
     target_compile_definitions(cjose_objects PRIVATE HAVE_OPENSSL_FEC_H=1)
 endif()
 if(MSVC)
-    target_compile_options(cjose_objects PRIVATE /W3 /WX)
+    target_compile_options(cjose_objects PRIVATE /W3 /WX /wd4267)
```
TheStormN, line 120: On Windows there are tons of warnings as errors about size truncation. However, fixing them will have to be a completely separate PR as the changes are a lot.

README.md
```diff
@@ -8,7 +8,7 @@ Implementation of JOSE for C/C++
 
 ### Build Tools ###
 
-* CMake (>= 3.23)
+* CMake (>= 3.22)
```
TheStormN, line 11: Fixed leftover from previous PR.

balthorium (approved):

Issue #142, opened 2026-09-02
roadmap and collaboration with OpenIDC/cjose


zandbelt:
@balthorium  @TheStormN 

We (@OpenIDC)  started a 0.6.2.x maintenance fork about 4.5 years ago because the project seemed to be abandoned since 2019, i.e. no commits since then. A number of Linux distro's started to carry our fork and some of them have backported our fixes over the years. Recently you have been updating the project again, incorporating fixes that we have been carrying for years, and adding CMake support on top of it. 

We would love to drop the maintenance fork for this branch again but then we'd need to make sure that all code in our fork is included here and we want to make sure that the project is actively maintained again going forward. (also we'd need to align on version numbers etc. ). 

Also: we started a [1.0.x](https://github.com/OpenIDC/cjose/tree/version-1.0.x) fork that uses OpenSSL 3.x primitives only and drops backwards compatibility for older (<3.0) OpenSSL versions. This cleans up a lot of `#ifdef`'d code and potential pitfalls and may be a good starting point as well for you.

Can you tell us what your plans are with this and can we perhaps work together?

FWIW: we are a bit frustrated about the lack of communication in the past - and recently again... - on this. we were forced to start the maintenance fork because of a lack of clarity and communication about cjose's future and we'd love to avoid it this time around.

Thanks

TheStormN:
Hello @zandbelt !

Thank you and all the @OpenIDC community for your support and patches over the years! My encounter with `cjose` is quite recent and thus the need for some maintenance came into need, especially about Windows DLL flavor of the library.

We are currently having internal discussions about the future library  maintenance and while there is a consensus about bringing `cjose` back to live, there are still certain details which have to be clarified. Once I have more information I will get back to you with a reply on this issue.

zandbelt:
excellent, thank you so much for your response; looking forward to what comes next; we are fine with any direction that you'll choose as long as the direction is clear so we can avoid any duplicate work/maintenance and/or conflicts

fluffy:
Hi @zandbelt - We had lost the main people originally supporting this project but trying to get it back on track. @TheStormN is going to help us get it going again. I have invited you to the Cisco org and hopefully you and @TheStormN can help figure out if we could get you as a committer to project and help merge the forks in. Thanks, Cullen

zandbelt:
[...]

PR #146, opened 2026-09-09
jwe: fix leaks, NULL dereferences and undefined behaviour on the JWE paths

zandbelt:
Part of the step 1 series (bug and undefined-behaviour fixes only, no public API change) discussed in #142, porting fixes carried in the OpenIDC/cjose `version-0.6.2.x` branch. This PR collects the memory-safety and crash fixes on the JWE paths; the RSA key-unwrap hardening follows in a separate PR.

### Changes

- `_cjose_jwe_encrypt_dat_aes_cbc()`: the `EVP_CIPHER_CTX` was leaked when the authentication-tag computation failed (`return false` instead of the `goto` cleanup). (OpenIDC/cjose@fb9abd5)
- ECDH-ES encrypt/decrypt leaked the empty `enc_key.raw` buffer that `cjose_jwe_import()` allocates for the encrypted-key segment. (OpenIDC/cjose@1b2af41)
- `_cjose_jwe_decrypt_ek_ecdh_es()` and `cjose_concatkdf_create_otherinfo()` dereferenced the optional `err` argument unconditionally, so an ECDH-ES encrypt or decrypt through the public API with `err == NULL` crashed. Regression test `test_cjose_jwe_ecdh_es_null_err` added. (OpenIDC/cjose@4e7e518, OpenIDC/cjose@6d7d79d)
- `cjose_jwe_decrypt_multi()`: check the CEK copy allocation, compare the CEKs of successive recipients in constant time (`cjose_const_memcmp`) instead of `memcmp`, and fail with `CJOSE_ERR_CRYPTO` when no recipient produced a CEK instead of calling `decrypt_dat` with a NULL key. The last item is #89 by @veselov, cherry-picked with its test to keep the series self-contained; happy to drop it here if you would rather merge #89 directly. (OpenIDC/cjose@54d4494, OpenIDC/cjose@2dbafde, OpenIDC/cjose@8b9416e)
- Wipe the decrypted plaintext held in `jwe->dat` on release and before reuse, so unauthenticated plaintext left behind by a failed AES-GCM tag check does not linger in freed memory. (OpenIDC/cjose@2d93399)
- `_cjose_jwe_malloc()`: tolerate `malloc(0)` returning NULL and skip `memset(NULL, 0, 0)`, which is undefined behaviour even with a zero length. (OpenIDC/cjose@ff74c49, OpenIDC/cjose@9238f5c)
- `_cjose_jwe_decrypt_dat_aes_cbc()`: compute the plaintext buffer size in `size_t`. `p_len + AES_BLOCK_SIZE` was evaluated in `int`, and the preceding guard only rejects `raw_len > INT_MAX`, so a ciphertext within 16 bytes of `INT_MAX` overflowed the addition. (OpenIDC/cjose@23e392a)
- Drop a duplicated `enc_iv` conversion in `_cjose_convert_to_base64()`. (OpenIDC/cjose@f84368a)

### Testing

Built with CMake (`-pedantic -Wall -Werror`) on Linux with GCC 15, OpenSSL 3.5.5 and Jansson 2.14; `check_cjose` passes. `valgrind --leak-check=full` on the jwe suite (`CK_FORK=no CK_RUN_SUITE=jwe`) reports 0 errors and 0 bytes lost.

### DIFF
--- a/src/concatkdf.c
+++ b/src/concatkdf.c
@@ -66,10 +66,16 @@ bool cjose_concatkdf_create_otherinfo(const char *alg,
     uint8_t *apu = NULL, *apv = NULL;
     size_t apuLen = 0, apvLen = 0;
 
-    memset(err, 0, sizeof(cjose_err));
+    // err is optional and may be NULL, so only dereference it when provided.
+    // cjose_header_get() records an error only for an invalid header/attr; for a
+    // valid hdr and the constant APU/APV attrs an absent field just yields NULL.
+    if (NULL != err)
+    {
+        memset(err, 0, sizeof(cjose_err));
+    }
     const char *apuStr = cjose_header_get(hdr, CJOSE_HDR_APU, err);
     const char *apvStr = cjose_header_get(hdr, CJOSE_HDR_APV, err);
-    if (CJOSE_ERR_NONE != err->code)
+    if (NULL != err && CJOSE_ERR_NONE != err->code)
     {
         return false;
     }
--- a/src/jwe.c
+++ b/src/jwe.c
@@ -2089,6 +2089,10 @@ uint8_t *cjose_jwe_decrypt_multi(cjose_jwe_t *jwe, cjose_key_locator key_locator
         {
             cek_len = jwe->cek_len;
             cek = cjose_get_alloc()(cek_len);
+            if (!cek) {
+               CJOSE_ERROR(err, CJOSE_ERR_NO_MEMORY);
+               return NULL;
+            }
             memcpy(cek, jwe->cek, cek_len);
         }
         else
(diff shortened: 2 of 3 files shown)


src/concatkdf.c
```diff
@@ -66,10 +66,16 @@ bool cjose_concatkdf_create_otherinfo(const char *alg,
...
+    {
+        memset(err, 0, sizeof(cjose_err));
+    }
     const char *apuStr = cjose_header_get(hdr, CJOSE_HDR_APU, err);
     const char *apvStr = cjose_header_get(hdr, CJOSE_HDR_APV, err);
-    if (CJOSE_ERR_NONE != err->code)
+    if (NULL != err && CJOSE_ERR_NONE != err->code)
     {
         return false;
     }
```
TheStormN: I think the return condition should remain the same, regardless if `err` is null or not. In case `err` null, the error will be ignored and code will continue with calling `cjose_base64url_decode()` when it will fail anyway.

TheStormN (approved):

PR #174, opened 2026-09-11
Eight consistency fixes across the library, tests, build and docs

zandbelt:
A sweep for places where the library, the tests, the two build systems and the docs do the same thing in two different ways, one commit per item, on top of #173's `cjose_const_memcmp` unification. Only the first commit changes what the library accepts; the rest are behaviour-preserving.

1. **jwe: require an empty encrypted key for dir and ECDH-ES.** RFC 7516 section 5.2 step 12. The AES key wrap, ECDH-ES+A*KW and RSA paths check the encrypted key for its exact expected size before handing it to OpenSSL; the `dir` and direct `ECDH-ES` paths never looked at it, so a JWE carrying one was decrypted as if it were absent. Now `CJOSE_ERR_INVALID_ARG`, with a regression test for both algorithms that fails on master.
2. **jwk: report a missing import string as `CJOSE_ERR_INVALID_ARG`.** `cjose_jwk_import(NULL, ...)` and a zero length returned NULL without touching `err`, unlike every other entry point. Regression test.
3. **jws: wipe the payload on release like the JWE plaintext.** `cjose_jwe_release` wipes the decrypted plaintext; `cjose_jws_release` wiped the digest and signature but freed the payload, its base64url encoding and the compact serialization with a plain dealloc.
4. **test: release library allocations with `cjose_get_dealloc`.** 22 places freed strings from `cjose_jwk_to_json` and `cjose_jwe_export_json`, plaintext from `cjose_jwe_decrypt` and buffers from `cjose_base64url_decode` with `free()`, which only works while the allocators are the defaults; the other 140 already used `cjose_get_dealloc()`.
5. **build: give the autotools test build the flags of the library and of CMake.** `test/Makefile.am` compiled the tests without `-Werror` and without `OPENSSL_API_COMPAT=0x10100000L`, both of which `src/Makefile.am` and CMake set. `Makefile.in` mirrored by hand.
6. **Prefix every static helper with `_cjose_` and its module.** jws.c and jwe.c already did; jwk.c had 28 helpers named after the key type alone, plus a handful in jwe.c, base64.c, concatkdf.c and a static `cjose_apply_allocs` in util.c that looked public. Mechanical rename.
7. **Add the copyright header to cjose.h**, the only file without one.
8. **README: list the supported algorithms, key types and serializations.** The additions of the recent releases were only discoverable through CHANGELOG.md and the headers.

Verified with CMake (`-pedantic -Wall -Werror`, RSA1_5 off and on), the full `check_cjose` run and valgrind on all suites, the `clang-format` target (no diff), an autotools `./configure && make && make check` confirming the test objects now compile with the new flags, and builds against OpenSSL 1.0.1u, 1.1.1w, 3.0.7 and 4.0.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,52 @@
 
 Implementation of JOSE for C/C++
 
+## Supported Algorithms ##
+
+JWS signing algorithms (`alg`):
+
+| Identifier | Algorithm | Requires |
+|------------|-----------|----------|
+| `HS256`, `HS384`, `HS512` | HMAC with SHA-2 | |
+| `RS256`, `RS384`, `RS512` | RSASSA-PKCS1-v1_5 with SHA-2 | |
+| `PS256`, `PS384`, `PS512` | RSASSA-PSS with SHA-2 | |
+| `ES256`, `ES384`, `ES512` | ECDSA with P-256, P-384 and P-521 | |
+| `ES256K` | ECDSA with secp256k1 | OpenSSL built with `secp256k1` |
+| `Ed25519`, `Ed448` | EdDSA (RFC 9864) | OpenSSL 1.1.1 |
+
+The polymorphic `EdDSA` identifier of RFC 8037, deprecated by RFC 9864, and
+`none` are not accepted.
+
+JWE key management algorithms (`alg`):
+
+| Identifier | Algorithm | Requires |
+|------------|-----------|----------|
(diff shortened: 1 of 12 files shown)


zandbelt:
@TheStormN perhaps you'd like to review the headers in all source files wrt. the copyright year(s)

TheStormN (approved):

PR #184, opened 2026-09-12
Add the A128GCMKW, A192GCMKW and A256GCMKW key encryption algorithms (RFC 7518 section 4.7)

zandbelt:
## Summary

Adds the AES GCM key wrapping algorithms `A128GCMKW`, `A192GCMKW` and `A256GCMKW` of RFC 7518 section 4.7. The CEK is encrypted with AES GCM under an `oct` key of exactly the size the `alg` names, with a random 96-bit IV and no additional authenticated data; the IV and the 128-bit authentication tag travel as the `iv` and `tag` header parameters of the recipient.

- **Parameter placement.** With one recipient the parameters go into the protected header, so the compact serialization carries them (the same choice as `epk` for ECDH-ES). With several recipients each recipient's `iv` and `tag` go into its own header, so `cjose_jwe_encrypt_multi` works with a different key encryption key per recipient. On decrypt they are read through the recipient, shared and protected headers in that order, which is where other implementations put them in the JSON serialization. The caller's header objects are never modified: the JWE already keeps a deep copy of the protected header and now takes one of a recipient header it writes to.
- **Checks before OpenSSL is called.** The key must be an `oct` key of the algorithm's size, the IV must be 12 octets and the tag 16, and the encrypted key must be exactly as long as the CEK the `enc` header dictates, so an attacker-controlled encrypted key can never be written past the CEK buffer. A failed tag verification wipes the unauthenticated CEK output.
- **API.** New macro-only constants `CJOSE_HDR_ALG_A128GCMKW`, `CJOSE_HDR_ALG_A192GCMKW`, `CJOSE_HDR_ALG_A256GCMKW`, `CJOSE_HDR_IV` and `CJOSE_HDR_TAG` (no compat symbols, no export-list changes, per the convention set by `CJOSE_HDR_ALG_ES256K`); `iv` and `tag` are accepted in `crit`. A README row.
- **Tests.** Round trips over all six content encryption algorithms; python-jwcrypto vectors (three compact serializations and a JSON serialization with the parameters in the recipient header); a two-recipient JSON round trip that also checks the caller's headers stay untouched; negative cases for the wrong key size and type, a tampered, missing, short or non-string `iv`/`tag`, and an encrypted key of the wrong length.

With this and the PBES2 family, which follows in a separate PR, cjose covers every algorithm of RFC 7518 (refs OpenIDC/cjose#28).

## Testing

- `-pedantic -Wall -Werror` build and the full `check_cjose` run (122 checks) with `CJOSE_ENABLE_RSA1_5` OFF and ON against OpenSSL 3.5.5.
- valgrind on the `jwe` suite: no leaks, no errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ JWE key management algorithms (`alg`):
 | `RSA-OAEP-256` | RSAES OAEP with SHA-256 and MGF1 with SHA-256 | |
 | `RSA1_5` | RSAES-PKCS1-v1_5 | build option `CJOSE_ENABLE_RSA1_5` |
 | `A128KW`, `A192KW`, `A256KW` | AES Key Wrap | |
+| `A128GCMKW`, `A192GCMKW`, `A256GCMKW` | AES GCM key wrapping | |
 | `dir` | direct use of a shared symmetric key | |
 | `ECDH-ES` | ECDH-ES direct key agreement | |
 | `ECDH-ES+A128KW`, `ECDH-ES+A192KW`, `ECDH-ES+A256KW` | ECDH-ES with AES Key Wrap | |
--- a/include/cjose/header.h
+++ b/include/cjose/header.h
@@ -42,6 +42,10 @@ extern "C" {
 #define CJOSE_HDR_APU "apu"
 #define CJOSE_HDR_APV "apv"
 
+/** For the AES GCM key wrapping algorithms, the initialization vector and the authentication tag of the encrypted key. */
+#define CJOSE_HDR_IV "iv"
+#define CJOSE_HDR_TAG "tag"
+
 /** The JWA algorithm attribute value for none. */
 #define CJOSE_HDR_ALG_NONE "none"
 
@@ -61,6 +65,11 @@ extern "C" {
 #define CJOSE_HDR_ALG_A128KW "A128KW"
 #define CJOSE_HDR_ALG_A192KW "A192KW"
 #define CJOSE_HDR_ALG_A256KW "A256KW"
+
+/** The JWE algorithm attribute value for A128GCMKW, A192GCMKW and A256GCMKW (AES GCM key wrapping, RFC 7518 section 4.7). */
+#define CJOSE_HDR_ALG_A128GCMKW "A128GCMKW"
+#define CJOSE_HDR_ALG_A192GCMKW "A192GCMKW"
+#define CJOSE_HDR_ALG_A256GCMKW "A256GCMKW"
 
 /** The JWE algorithm attribute value for ECDH-ES with A128KW, A192KW or A256KW key wrapping. */
 #define CJOSE_HDR_ALG_ECDH_ES_A128KW "ECDH-ES+A128KW"
--- a/src/jwe.c
+++ b/src/jwe.c
@@ -39,6 +39,12 @@ static bool _cjose_jwe_encrypt_ek_aes_kw(_jwe_int_recipient_t *recipient, cjose_
 static bool _cjose_jwe_decrypt_ek_aes_kw(_jwe_int_recipient_t *recipient, cjose_jwe_t *jwe, const cjose_jwk_t *jwk, cjose_err *err);
 
 static bool
+_cjose_jwe_encrypt_ek_aes_gcm_kw(_jwe_int_recipient_t *recipient, cjose_jwe_t *jwe, const cjose_jwk_t *jwk, cjose_err *err);
+
+static bool
+_cjose_jwe_decrypt_ek_aes_gcm_kw(_jwe_int_recipient_t *recipient, cjose_jwe_t *jwe, const cjose_jwk_t *jwk, cjose_err *err);
+
+static bool
 _cjose_jwe_encrypt_ek_rsa_oaep(_jwe_int_recipient_t *recipient, cjose_jwe_t *jwe, const cjose_jwk_t *jwk, cjose_err *err);
 
 static bool
@@ -318,6 +324,52 @@ static const char *_cjose_jwe_get_from_headers(cjose_header_t *protected_header,
     return NULL;
 }
 
+// like _cjose_jwe_get_from_headers, but returns the JSON value so that a
+// parameter can be checked for its expected type (borrowed reference)
+static json_t *_cjose_jwe_get_json_from_headers(cjose_header_t *protected_header,
+                                                cjose_header_t *unprotected_header,
+                                                cjose_header_t *personal_header,
+                                                const char *key)
+{
+    cjose_header_t *headers[] = { personal_header, unprotected_header, protected_header };
+    for (int i = 0; i < 3; i++)
+    {
+        if (NULL == headers[i])
+        {
+            continue;
+        }
+        json_t *obj = json_object_get((json_t *)headers[i], key);
+        if (NULL != obj)
+        {
+            return obj;
+        }
+    }
+    return NULL;
+}
+
+// the header object a per-recipient parameter (like the "iv" and "tag" of the
+// AES GCM key wrapping algorithms) is written to when encrypting: the protected
+// header for a single recipient, so that the compact serialization carries it,
+// and the recipient's own header otherwise. The caller's header objects are
+// never modified: the JWE holds a deep copy of the protected header already and
+// gets one of the recipient header here.
+static json_t *_cjose_jwe_recipient_param_target(cjose_jwe_t *jwe, _jwe_int_recipient_t *recipient, cjose_err *err)
+{
+    if (jwe->to_count < 2)
+    {
+        return jwe->hdr;
+    }
+    json_t *copy = (NULL == recipient->unprotected) ? json_object() : json_deep_copy(recipient->unprotected);
+    if (NULL == copy)
+    {
+        CJOSE_ERROR(err, CJOSE_ERR_NO_MEMORY);
+        return NULL;
+    }
+    json_decref(recipient->unprotected);
+    recipient->unprotected = copy;
+    return copy;
+}
+
 static bool _cjose_jwe_validate_enc(cjose_jwe_t *jwe, cjose_header_t *protected_header, cjose_err *err)
 {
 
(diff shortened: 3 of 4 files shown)


TheStormN (changes requested):
There seems to be some issues with this PR which needs to be corrected. I will review the other pending PRs once this one is merged.

1. [P1] Critical-header support is too broad — [src/jwe.c:416](https://github.com/cisco/cjose/blob/18ba61c6d4e77cdb2a16b9fb500d79a0366ad766/src/jwe.c#L416)
   Adding "iv" and "tag" to the global supported_crit_headers list makes them accepted as critical parameters for every JWE algorithm, including RSA-OAEP and A128KW, where the implementation does not process them. A JWE with crit:["iv"] can therefore be accepted while ignoring the critical parameter.
   RFC 7516 requires critical parameters to be understood and processed, while RFC 7518 defines iv/tag as additional parameters specifically for the A*GCMKW algorithms. Validate these parameters conditionally based on alg, or reject them for other algorithms. ([RFC 7518 §4.7](https://www.rfc-editor.org/rfc/rfc7518.html#section-4.7), [RFC 7516 §4.1.13](https://www.rfc-editor.org/rfc/rfc7516.html#section-4.1.13))
2. [P2] Duplicate iv/tag header parameters are accepted and can be emitted — [src/jwe.c:979](https://github.com/cisco/cjose/blob/18ba61c6d4e77cdb2a16b9fb500d79a0366ad766/src/jwe.c#L979)
   _cjose_jwe_get_json_from_headers() silently selects the per-recipient value over shared/protected values. In JSON serialization, however, header names across these locations must be disjoint. The encryption path can also create duplicates if callers already supplied iv or tag in another header object.
   Reject duplicate names during validation rather than relying on precedence. ([RFC 7516 §5.2](https://www.rfc-editor.org/rfc/rfc7516.html#section-5.2), [RFC 7516 §7.2.1](https://www.rfc-editor.org/rfc/rfc7516.html#section-7.2.1))

zandbelt:
Thanks — both addressed, in two commits on top of the branch rather than a rewrite, so the review stays readable.

**1. Critical headers (a3ec65b).** The supported set is now chosen after `alg` is resolved: `iv` and `tag` are understood as critical parameters for A128GCMKW/A192GCMKW/A256GCMKW and refused for every other algorithm, on encryption and on import alike. The test covers both directions — accepted for the GCM key wrapping algorithms, `CJOSE_ERR_INVALID_ARG` for A128KW and dir, and a JWE whose protected header marks `iv` critical is refused at `cjose_jwe_import` unless its `alg` is one of the three.

Worth flagging that the same pattern is already on master, independently of this PR: `epk`, `apu` and `apv` are ECDH-ES parameters that the global list accepts for every algorithm, so `{"alg":"A128KW","crit":["epk"]}` is accepted today and `epk` is ignored. I left that alone here to keep this PR to its subject. Strictly, RFC 7515 section 4.1.11 goes further than scoping: producers must not list *any* name defined by the JWS/JWE/JWA specifications in `crit`, which covers `alg`, `enc`, `cty`, `epk`, `apu`, `apv`, `iv` and `tag` — the whole list. Refusing them all would be the conformant reading, but it changes behaviour for anything already producing those, so I would rather do it as its own PR if you want it; say the word.

**2. Duplicate parameters (9f1c961).** The encryption path no longer creates them: if `iv` or `tag` is already present in the protected header, the shared unprotected header or the recipient's header, encryption fails with `CJOSE_ERR_INVALID_ARG` instead of writing a second copy. This was a real defect and not only a theoretical one — with a single recipient and a shared unprotected header carrying `iv`, the old code emitted the parameter in both the protected and the shared header, and the resulting JWE could not be decrypted by cjose itself, because the lookup preferred the shared copy and the GCM tag check then failed. The test covers all three header locations for both parameters.

The other half, rejecting duplicate names when reading a JWE, I did not do here, because it is not specific to this PR and not specific to `iv`/`tag`: `_cjose_jwe_get_from_headers()` on master resolves `alg`, `enc` and `kid` with exactly the same precedence, so a JSON JWE that repeats any of those across the three locations is accepted today and the per-recipient copy wins. The sharper form of your point is that the winning copy is the *unprotected* one, which is not integrity protected, while the protected header is the AEAD's additional data — for `alg` that is the shape of an algorithm-confusion bug, even though it does not seem exploitable in practice, since a substituted algorithm fails either the key-type check in `_cjose_jwe_validate_decrypt_key()` or the content authentication. Enforcing the disjointness of RFC 7516 section 7.2.1 across the three locations touches every algorithm and both import paths, so I would prefer it as a separate PR, with its own tests, rather than smuggled in here. Happy to open it right after this one.

Testing: full `check_cjose` with RSA1_5 off and on, valgrind clean on the `jwe` suite, `clang-format` target produces no diff, and both new tests fail on the branch before these commits.

TheStormN (changes requested):
So on to the next pass. Btw, while I do understand that other algorithms implementations suffer from similar issues. Perhaps v1 is now good time to address all of that(in a separate PR ofc).

[P1] crit is still accepted in unprotected headers — [src/jwe.c:445-447](https://github.com/cisco/cjose/blob/9f1c9615efc0125956da3e3bd2265db9ea00ea29/src/jwe.c#L445-L447)
The algorithm-specific critical-header scope is now correct. However, validation still checks crit in the shared and per-recipient unprotected headers. RFC 7516 requires crit to be integrity protected, meaning it must appear only in the protected header.
For example, an A128GCMKW JWE with crit:["iv"] in a recipient’s unprotected header is accepted. Validate crit only in protected_header; reject it elsewhere. ([RFC 7516 §4.1.13](https://www.rfc-editor.org/rfc/rfc7516.html#section-4.1.13))

[P2] Imported JSON JWEs can still contain duplicate iv/tag names — [src/jwe.c:1006-1009](https://github.com/cisco/cjose/blob/9f1c9615efc0125956da3e3bd2265db9ea00ea29/src/jwe.c#L1006-L1009)
The new check correctly prevents the encryption API from emitting duplicates supplied by callers. But decryption still uses first-match precedence across recipient, shared, and protected headers. A manually supplied JSON JWE containing the same iv or tag in multiple locations is accepted instead of rejected, contrary to RFC 7516’s requirement that header names be disjoint.
Add duplicate-name validation during JSON import or before GCM-KW decryption. ([RFC 7516 §5.2](https://www.rfc-editor.org/rfc/rfc7516.html#section-5.2))

zandbelt:
Both are real, and both are already fixed — in #188, which I opened shortly before this review landed, so it will not have been on your radar. It is the separate PR your opening line asks for: it takes the same two rules across all algorithms rather than only the AES GCM key wrapping ones.

**P1** is exactly what #188 does with `crit`: the list is validated only in the protected header and refused anywhere else, since it has to be integrity protected. **P2** is the RFC 7516 section 7.2.1 disjointness check it adds, which runs per recipient on import, so a JSON JWE that repeats a name in two of the three locations is rejected rather than resolved by precedence.

Your two examples, built as a two-recipient A128GCMKW/A256GCMKW JWE and imported:

| case | this PR (9f1c961) | with #188 |
|---|---|---|
| `crit:["iv"]` in a recipient's unprotected header | accepted | rejected |
| the same `iv` in the shared and the recipient header | accepted | rejected |
| the same `tag` in the shared and the recipient header | accepted | rejected |
| cjose's own output, untouched | accepted | accepted |

One thing worth recording: with one narrow exception neither is a regression from this PR. On master today, `crit` in an unprotected header is accepted — `crit:["alg"]` in a recipient header imports fine — and so are duplicate names across the three locations. The exception is that this PR made `iv` and `tag` listable in `crit` at all, so your P1 example only became reachable here, which is the argument for fixing the rule generally instead of patching it for one algorithm family.

#188 also covers the rest of what you point at: `epk`, `apu` and `apv` scoped to ECDH-ES the way `iv` and `tag` are scoped here, the remaining RFC 7515 section 4.1.11 rules — no duplicate names in the list, a listed name must occur in the JOSE header — for JWE and JWS alike, and the disjointness check on the encryption path as well as on import. Its description also spells out an alternative, stricter reading of section 4.1.11, in case you would rather refuse `crit` outright.

On merge order, your call:

1. Review #188 and merge that. It is stacked on this branch, so its diff carries these three commits and one merge lands everything.
2. Keep this PR self-contained. I move the `crit` placement and the disjointness commits down into this branch and rebase #188 to the ECDH-ES scoping and the remaining producer rules.

I would suggest the first, since both fixes are cross-algorithm and read better that way, but the second is a short rebase if you prefer this PR to stand on its own.

zandbelt:
Rebased to keep this PR to the algorithms, as you asked. It is two commits now:

- the A128GCMKW, A192GCMKW and A256GCMKW implementation, unchanged;
- refusing a caller-supplied `iv` or `tag` when encrypting, which is the fix for the second point of your first review and belongs with the algorithms, since those two parameters are what the key wrapping produces.

Gone from here is the commit that scoped `iv` and `tag` as critical parameters, together with its test. #188 refuses a `crit` list outright, so that mechanism disappears there anyway and it should not land here first. The `_cjose_jwe_alg_is_aes_gcm_kw()` helper it introduced went with it; #188 brings it back, since that is where it is now needed.

Nothing else changed: same algorithms, same tests for them, same interoperability vectors. Full `check_cjose` with RSA1_5 off and on, valgrind clean, `clang-format` no diff.

zandbelt:
Closing this: it is already in master. #188 was stacked on this branch, so merging it as 3c153f7 carried these two commits in with it — the A128GCMKW, A192GCMKW and A256GCMKW implementation, the header constants, the README row, the interoperability vectors and the refusal of a caller-supplied `iv` or `tag` are all there.

I checked rather than assumed: rebasing this branch onto master leaves no commits and no difference from it, since both sides add the same code, which is also why GitHub calls the PR conflicting. So there is nothing left here to merge, and the "what moved where" note I left earlier was overtaken by the merge, which happened a few minutes before I rebased.

Thanks for the two review rounds on this one — the critical-header scope and the duplicate parameters both turned out to be worth more than the algorithms they were found in.

TheStormN:
Thanks @zandbelt It's good that these issues were discovered. I will take on your other PRs when you sync them with the `master`, just tag me when ready.

PR #188, opened 2026-09-12
Refuse a "crit" header and enforce the disjointness of the JWE header locations

zandbelt:
## Breaking change, please read first

**A JWE or JWS whose header carries a `crit` list is refused, on production and on consumption alike.** Until now a list was accepted as long as it named parameters cjose processes — `alg`, `enc` and `cty` for both, plus `epk`, `apu` and `apv` for a JWE — and `iv` and `tag` were added to that set by #184. `cjose_jws_sign`, `cjose_jwe_encrypt*`, `cjose_jws_import` and `cjose_jwe_import*` now fail with `CJOSE_ERR_INVALID_ARG` for any of them.

RFC 7515 section 4.1.11 is the reason: `crit` names extensions to the JOSE specifications that the recipient has to understand and process, a recipient must reject a list naming anything it does not support, and a producer must not list a name the specifications or JWA define, which a recipient may reject in turn. cjose implements no extension, so every name a list can carry falls under one of those two, and the conformant behaviour collapses to refusing the list.

Consequences worth being deliberate about:

- A caller that sets `crit` on a header it passes to cjose now gets an error instead of a token. There is no API change and no new symbol; only the behaviour.
- A token from a producer that emits `crit` naming registered parameters is refused where it used to be accepted. Such a producer is already outside RFC 7515 section 4.1.11, but it will be a visible change for whoever meets one.
- It supersedes the scoping of `iv` and `tag` in #184: that commit is in this branch and this one removes the mechanism it added. Happy to drop it from #184 instead, if you would rather that PR not carry a mechanism that disappears here.
- For the release notes, whenever you write them: *a JOSE header that carries a "crit" parameter is refused; a list naming parameters cjose processes, such as "alg", "enc" or "cty", used to be accepted.*

### Disjoint header names (RFC 7516 section 7.2.1)

The member names of the JWE protected header, the shared unprotected header and a per-recipient unprotected header must be disjoint. They were not checked, and `_cjose_jwe_get_from_headers()` resolves a name by precedence — per-recipient, then shared, then protected — so a JWE that repeated a name was accepted and the *unprotected*, unauthenticated copy won. On master a JWE whose protected header says `{"alg":"A128KW"}` and which an attacker gives a per-recipient `{"alg":"dir"}` imports fine and the substituted algorithm drives the decryption; it is now refused. For `alg` that is the shape of an algorithm-confusion bug, though it does not appear to be exploitable, since a substituted algorithm fails either the key type check in `_cjose_jwe_validate_decrypt_key()` or the content authentication.

The lookup order is deliberately left as it is. Reversing it to prefer the protected header would be wrong, because the per-recipient header is the legitimate place for `alg` in a JSON serialization and valid multi-recipient JWEs would break. The invariant that makes the precedence safe is that exactly one location carries a given name, which is what the disjointness check establishes, and it runs before any header value is acted upon: the compact path validates the recipient before `_cjose_jwe_validate_enc()`, and the JSON path validates every recipient before it. A further guard inside the lookup helper, asserting a single source, would therefore be unreachable code, so there is none.

### The headers a JWE leaves the encryption with

Validation ran on the caller's headers only, so it never saw the parameters the algorithms write themselves and the encryption could produce a JWE that cjose's own import refuses: a caller-supplied `epk` in the shared or a per-recipient header survived next to the generated one. The assembled headers are therefore checked again once every recipient has been encrypted to, which gives the property that was missing — what `cjose_jwe_encrypt*` returns is a JWE that `cjose_jwe_import*` accepts.

A parameter the algorithm produces itself is also refused up front when it comes from the caller, as #184 does for `iv` and `tag`, and that now covers `epk`, which was silently replaced by the generated key when the caller set it in the protected header. The three uses share one helper.

One related fix this needed: `cjose_jwe_encrypt_multi_iv()` passed the recipient's own header to `_cjose_jwe_validate_alg()` in the place of the shared unprotected header, so the shared header was invisible to validation and, as a side effect, `alg` could not be resolved from it when encrypting, while the import path did resolve it there. The call now passes the shared header, which makes the two paths symmetric.

## Testing

- `check_cjose` in full, with RSA1_5 off and on, and valgrind clean on the whole run; the `clang-format` target produces no diff.
- The `crit` refusal is exercised for four algorithms against four different lists, in the protected header and in an unprotected one, when encrypting and when importing, for JWE and for JWS.
- Unit tests for the refusal and for the disjointness check in the `header` suite, and JWE regression tests for the generated parameters and for the round trip.
- Verified against the tip of #184 that the behaviours change: `crit` in every shape above, a JSON serialization repeating `enc` across the protected and a recipient header, the `alg` shadowing above, a caller-supplied `epk`, and an ECDH-ES JWE that the encryption emitted and the import refused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ JWE key management algorithms (`alg`):
 | `RSA-OAEP-256` | RSAES OAEP with SHA-256 and MGF1 with SHA-256 | |
 | `RSA1_5` | RSAES-PKCS1-v1_5 | build option `CJOSE_ENABLE_RSA1_5` |
 | `A128KW`, `A192KW`, `A256KW` | AES Key Wrap | |
+| `A128GCMKW`, `A192GCMKW`, `A256GCMKW` | AES GCM key wrapping | |
 | `dir` | direct use of a shared symmetric key | |
 | `ECDH-ES` | ECDH-ES direct key agreement | |
 | `ECDH-ES+A128KW`, `ECDH-ES+A192KW`, `ECDH-ES+A256KW` | ECDH-ES with AES Key Wrap | |
--- a/include/cjose/header.h
+++ b/include/cjose/header.h
@@ -42,6 +42,10 @@ extern "C" {
 #define CJOSE_HDR_APU "apu"
 #define CJOSE_HDR_APV "apv"
 
+/** For the AES GCM key wrapping algorithms, the initialization vector and the authentication tag of the encrypted key. */
+#define CJOSE_HDR_IV "iv"
+#define CJOSE_HDR_TAG "tag"
+
 /** The JWA algorithm attribute value for none. */
 #define CJOSE_HDR_ALG_NONE "none"
 
@@ -61,6 +65,11 @@ extern "C" {
 #define CJOSE_HDR_ALG_A128KW "A128KW"
 #define CJOSE_HDR_ALG_A192KW "A192KW"
 #define CJOSE_HDR_ALG_A256KW "A256KW"
+
+/** The JWE algorithm attribute value for A128GCMKW, A192GCMKW and A256GCMKW (AES GCM key wrapping, RFC 7518 section 4.7). */
+#define CJOSE_HDR_ALG_A128GCMKW "A128GCMKW"
+#define CJOSE_HDR_ALG_A192GCMKW "A192GCMKW"
+#define CJOSE_HDR_ALG_A256GCMKW "A256GCMKW"
 
 /** The JWE algorithm attribute value for ECDH-ES with A128KW, A192KW or A256KW key wrapping. */
 #define CJOSE_HDR_ALG_ECDH_ES_A128KW "ECDH-ES+A128KW"
--- a/src/header.c
+++ b/src/header.c
@@ -14,14 +14,31 @@
 static const char *CJOSE_HDR_CRIT = "crit";
 
 ////////////////////////////////////////////////////////////////////////////////
-bool _cjose_header_validate_crit(cjose_header_t *header, const char *const *supported, size_t supported_len, cjose_err *err)
+bool _cjose_header_validate_crit(
+    cjose_header_t *const *headers, size_t headers_len, const char *const *supported, size_t supported_len, cjose_err *err)
 {
-    if (NULL == header)
+    if (NULL == headers || 0 == headers_len)
     {
         return true;
     }
 
-    json_t *crit = json_object_get((json_t *)header, CJOSE_HDR_CRIT);
+    // the list has to be integrity protected, so it may only appear in the
+    // protected header (RFC 7515 section 4.1.11)
+    for (size_t i = 1; i < headers_len; i++)
+    {
+        if (NULL != headers[i] && NULL != json_object_get((json_t *)headers[i], CJOSE_HDR_CRIT))
+        {
+            CJOSE_ERROR(err, CJOSE_ERR_INVALID_ARG);
+            return false;
+        }
+    }
+
+    if (NULL == headers[0])
+    {
+        return true;
+    }
+
+    json_t *crit = json_object_get((json_t *)headers[0], CJOSE_HDR_CRIT);
     if (NULL == crit)
     {
         return true;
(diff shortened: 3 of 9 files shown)


TheStormN (changes requested):
So about the P2 issue as we already agreed, it will be in a separate PR. However, the other two P1 issues seems to be real.

Review findings:
1. [P1] Validate crit parameter presence during encryption
   [src/jwe.c:470](https://github.com/cisco/cjose/blob/bc1dbb2257f6c4efb1fd021db3dd5c5730c927ce/src/jwe.c#L470) validates supported names but not their presence. For example, ECDH-ES accepts crit:["apu"] or crit:["apv"] even when those parameters are absent; the resulting JWE is exported but rejected on re-import. Validate the completed header before returning, or reject missing names during encryption.
2. [P1] Header disjointness is checked before ECDH adds epk
   [src/jwe.c:442](https://github.com/cisco/cjose/blob/bc1dbb2257f6c4efb1fd021db3dd5c5730c927ce/src/jwe.c#L442) checks caller-supplied headers, but ECDH later adds epk to the protected header. If epk exists in a shared or recipient header, encryption emits duplicate header names and JSON import rejects the result. Check for collisions with generated parameters or validate after generation.
3. [P2] crit allowlist contradicts RFC rules
   [src/jwe.c:434-436](https://github.com/cisco/cjose/blob/bc1dbb2257f6c4efb1fd021db3dd5c5730c927ce/src/jwe.c#L434-L436) allows registered parameters (alg, enc, cty, epk, apu, apv, iv, tag) in crit. RFC 7515 §4.1.11 says producers must not list parameters defined by JWS/JWA; RFC 7516 applies the same rule to JWE. The new tests currently encode non-conforming behavior.

zandbelt:
Both P1 findings were real, and both came down to the same thing: the encryption validated the caller's headers and never looked at them again, so it could emit a JWE that its own import refuses. Fixed in the commit above, and thank you — the round-trip framing is what makes it obvious.

What the fix does is what you suggest in the second finding: after every recipient's key encryption has run, so after the algorithm has written the parameters it produces, the assembled headers are validated once more, per recipient, for disjointness and for the presence of every critical parameter. That covers both findings at once, and it gives the property that was missing: what `cjose_jwe_encrypt*` returns is a JWE that `cjose_jwe_import*` accepts.

On top of that, a parameter the algorithm produces itself is refused when it comes from the caller, the way #184 does for `iv` and `tag`. That now includes `epk`, which was not only ending up in two locations but was being silently replaced by the generated one when the caller put it in the protected header. The three uses share one helper.

Before and after, with your two examples:

| case | before | after |
|---|---|---|
| ECDH-ES `crit:["apu"]` with no `apu` | encrypt accepted, re-import rejected | encrypt rejected |
| caller `epk` in a recipient header | encrypt accepted, re-import rejected | encrypt rejected |
| caller `epk` in the shared or the protected header | accepted, silently replaced | encrypt rejected |
| ECDH-ES `crit:["epk"]`, `crit:["apu"]` with `apu` present | accepted | accepted, and re-imports |

The test that encoded the behaviour you flagged is corrected rather than removed: `crit:["apu"]` is now exercised with `apu` actually in the header, `crit:["epk"]` stays as it is because the key agreement writes `epk`, and both assert the round trip. Both new assertions fail without the fix.

About the third point, the `crit` allowlist against RFC 7515 section 4.1.11: happy to do that, and worth deciding before I write it, because the two are not independent. If a `crit` list is refused outright — which is where section 4.1.11 lands for a library that implements no extensions, since every possible entry is then either forbidden to the producer or unknown to the recipient — then the per-algorithm lists, the duplicate-name rule and the presence check in this PR all become dead code, and about half of what is here disappears again. Only the disjointness work survives.

So, your call, and either is fine by me:

1. I collapse this PR into the strict version now: `crit` refused wherever it appears, the allowlist and the scoping deleted, disjointness and the post-generation validation kept. Smaller than what is here today, and it supersedes the `crit` scoping of #184 as well.
2. This PR stays as it is and the strict change follows as its own PR, which then removes the machinery.

I lean towards the first, since it avoids merging a mechanism that the next PR deletes, but it does refuse `crit` lists that some producers emit today, so it is a behaviour change worth being deliberate about. Say which and I will push it.

TheStormN:
Ok lets go with option 1 and use this PR.

zandbelt:
Done — pushed as `5cf3a8d`, and the PR title and description are rewritten around it, since this is now a behaviour change that deserves to be the first thing a reader sees rather than a detail in the middle.

**A JWE or JWS whose header carries a `crit` list is refused, on production and on consumption alike.** What used to be accepted was a list naming parameters cjose processes — `alg`, `enc` and `cty` for both, plus `epk`, `apu` and `apv` for a JWE, and `iv` and `tag` since #184. All of those now fail with `CJOSE_ERR_INVALID_ARG`. No API change, no new symbol, only the behaviour.

What that removed: the per-algorithm sets of accepted critical parameters, the scoping of `iv`/`tag` and `epk`/`apu`/`apv` that went with them, and the rules on the contents of a list, which are unreachable once no list is accepted. The `crit` half of this PR is about a third of what it was. The disjointness work and the validation of the assembled headers are untouched, since neither has anything to do with `crit`.

The refusal is tested for four algorithms against four different lists, in the protected header and in an unprotected one, when encrypting and when importing, for JWE and for JWS. Full `check_cjose` with RSA1_5 off and on, valgrind clean, `clang-format` no diff.

Two things to decide, whenever it suits you:

- This supersedes the `crit` scoping of #184. That commit is in this branch and this change removes the mechanism it introduced, so #184 would merge a mechanism that disappears one PR later. I can drop it from #184 if you would rather keep that PR to the algorithms themselves; it is a rebase and the AES GCM key wrapping work is unaffected.
- The release-note line is yours to word, but for what it is worth: *a JOSE header that carries a "crit" parameter is refused; a list naming parameters cjose processes, such as "alg", "enc" or "cty", used to be accepted.* This branch predates the 1.0.0 section in CHANGELOG.md, so I have not touched it.

TheStormN (approved):

TheStormN:
Regarding your ask - Yes. lets keep #184 to the algorithms themself.

zandbelt:
Rebased onto the reduced #184 and squashed into a single commit.

Two reasons for the squash. The branch's history added the per-algorithm `crit` allowlist and then deleted it again two commits later, which is a confusing read for anyone but me; and with the scoping commit gone from #184, that history no longer applied cleanly. The commit that is here now is the net change: refuse a `crit` header, enforce the disjointness of the three JWE header locations, validate the assembled headers once the algorithms have written the parameters they produce, and refuse those parameters when they come from the caller.

**The tree is byte-identical to what you reviewed at `5cf3a8d`** — I checked file by file, not just the test results. Nothing you have already read changes; only the history leading to it, and what the diff shows on top of #184: `_cjose_jwe_alg_is_aes_gcm_kw()` now appears here, because the commit that introduced it left #184.

Still 128 checks with RSA1_5 off and 127 with it on, valgrind clean over the whole run, `clang-format` no diff.

PR #193, opened 2026-09-13
Refuse a private member whose value is zero

zandbelt:
## Summary

Third turn at one rule, so let me state it exactly: an RSA private member that is present in the JWK but carries **no usable value** must fail the import, instead of being read as absent and yielding a public key.

- #189 tested the decoded buffer pointer. Base64url padding on its own decodes to zero octets with a non-NULL buffer, so `"d": "=="` passed.
- #192 added the decoded length. A value of `"AA"` decodes to **one octet that is zero** — the integer 0 — with a length of 1, so it passed too.
- This one tests what the rule always meant: the decoded value is zero.

What it costs a caller on master today: a key whose `d`, `p`, `q`, `dp`, `dq` or `qi` is zero imports, and `cjose_jwk_to_json(key, true)` then writes the member back as `""`, which cjose itself refuses to read in again. A key that cannot survive its own export is not one to accept.

A value that merely *contains* zero octets is an ordinary number and keeps importing — `"AQA"` is `0x01 0x00` and `"AAE"` is `0x00 0x01`, one for each end of the scan, both asserted.

### What this does not do

It does not make an accepted key a *usable* one. Any other wrong `d` — `"AQ"`, say — still imports, still signs, and still produces a JWS that fails its own verification. Refusing that means checking the key for consistency rather than checking one member for a value, which is a change of a different size; I would rather raise it separately than grow this one.

Two neighbouring gaps I found while working on this, both older than these PRs and neither addressed here — happy to open issues if you want them tracked:

- `p`, `q`, `dp`, `dq` and `qi` with real values but no `d` are dropped silently and the key becomes public, so `{n, e, p}` is accepted where RFC 7518 section 6.3.2 requires the private members to appear together.
- The zero-valued-member behaviour above exists for EC and OKP keys too, though their fixed expected length already refuses the shapes that reach it.

## Testing

- The negative matrix now runs against a **complete private key** with one member zeroed, which is where the harm actually shows, as well as against a public key with a member appended. All six members, four valueless shapes.
- The accepted key is asserted to survive the round trip its own export implies.
- Verified failing on master: `a d of "AA" was accepted`.
- Full `check_cjose` with RSA1_5 off and on, `clang-format` produces no diff.

Reviewed by an independent agent before submitting, per your request on #185: it confirmed the predicate partitions every shape `cjose_base64url_decode` can return, found no false rejection over a 233-key corpus (random RSA at three sizes, with and without CRT parameters, RFC 7517 A.2, RFC 7515 A.2, each round-tripped), and is what prompted the stronger test and the scope note above.


TheStormN (commented):
While part of this review also applies to the other PR. Perhaps it would be simpler to be fixed here?

The PBES2 construction itself is correct: algorithm-to-digest/key-size mapping, salt composition, PBKDF2, AES-KW, and CEK-length validation all align with RFC 7518.
Findings:
1. P2 — p2c rejects RFC-valid values
   The code rejects counts below 1000 at [src/jwe.c:1339](https://github.com/cisco/cjose/blob/d9a4179880c4d2e871d9cfc2fa8dd9ca8fbf2c0d/src/jwe.c#L1339). RFC 7518 requires a positive JSON integer; 1000 is only recommended, not mandatory. Thus valid JWEs with p2c from 1–999 cannot be decrypted. See [RFC 7518 §4.8.1.2](https://www.rfc-editor.org/rfc/rfc7518.html#section-4.8.1.2).
2. P2 — undocumented upper bounds reduce RFC interoperability
   The hard limits of 1024 salt bytes and 1,000,000 iterations at [jwe_int.h:98-103](https://github.com/cisco/cjose/blob/d9a4179880c4d2e871d9cfc2fa8dd9ca8fbf2c0d/src/include/jwe_int.h#L98-L103) reject values for which RFC 7518 specifies no maximum. DoS protection is reasonable, but this should be documented as a library policy and ideally made configurable through a supported public option.
3. P2 — noncanonical/padded p2s can be emitted
   Caller-supplied p2s is decoded but preserved verbatim at [src/jwe.c:1235-1262](https://github.com/cisco/cjose/blob/d9a4179880c4d2e871d9cfc2fa8dd9ca8fbf2c0d/src/jwe.c#L1235-L1262). The decoder accepts = and ignores trailing data at [base64.c:99-102](https://github.com/cisco/cjose/blob/d9a4179880c4d2e871d9cfc2fa8dd9ca8fbf2c0d/src/base64.c#L99-L102), so encryption can produce p2s values such as AAAAAAAAAAA=. RFC base64url encoding excludes padding and extra characters; reject or canonicalize supplied values. See [RFC 7515 §2](https://www.rfc-editor.org/rfc/rfc7515.html#section-2).
Overall, I would request changes for the RFC conformance/interoperability points above; the cryptographic implementation otherwise looks sound.

zandbelt:
To answer the question directly: it cannot be fixed here, since this branch carries none of that code. #193 is one commit over `src/jwk.c` and `test/check_jwk.c`, so all three findings go to #185, and they are answered there.

Where they stand, since two of them were already dealt with by the time you wrote this (the links are to d9a4179, two commits behind #185's head at that point):

- **Padded `p2s` can be emitted.** Gone. A caller-supplied salt input is refused outright now, so the encrypt path has nothing verbatim to preserve. RFC 7518 section 4.8.1.1 wants a new random salt input per encryption anyway, which was your first finding on #185.
- **Undocumented upper bounds.** Documented as library policy in `jwe_int.h` and in the README. The "supported public option" half was a fair hit, and is now done properly: every PBES2 bound is a CMake option next to `CJOSE_ENABLE_RSA1_5`, with a row each in the README build-options table.
- **`p2c` rejects RFC-valid values.** You are right, and this one was still open. The floor now applies only to what cjose writes: it will not produce a JWE below the 1000 the RFC recommends, and it reads any positive count up to the ceiling, which is section 4.8.1.2 exactly. `CJOSE_JWE_PBES2_MIN_ACCEPTED_ITERATIONS` is there for a deployment that would rather fail closed on a weak count. There is a jwcrypto vector at `p2c=100` in the suite that fails without the change.

Thanks for pushing on that last one twice; the asymmetry is obviously right in hindsight. Refusing to read a low count never made the producer's token any stronger, and the ceiling is what actually protects the recipient.

This PR itself is unreviewed so far, whenever you get to it.

TheStormN (approved):

PR #195, opened 2026-09-13
A 0.8.x maintenance release: six fixes backported from the 1.0 line

### LINKED ISSUE #194: Propose a 0.8.x maintenance line for users on OpenSSL 1.x
Master is the 1.0 line now: OpenSSL 3.0 only, no autotools, `libcjose.so.1`, and `crit` refused outright. That is the right place for the library to be, but it leaves anyone on OpenSSL 1.x without a route to the fixes that landed alongside those changes, and 0.8.0 declares `find_package(OpenSSL 1.0.1)`.

So: would you like a 0.8.x maintenance line? I have prepared one to make the question concrete rather than theoretical — **[OpenIDC/cjose `version-0.8.x`](https://github.com/OpenIDC/cjose/commits/version-0.8.x)**, seven commits off the `0.8.0` tag. If you want it, create `version-0.8.x` here and I will open it as a PR against that branch; if you would rather everyone moved to 1.0, close this and nothing is lost.

### What is wrong with 0.8.0

Each of these I reproduced against a build of the `0.8.0` tag itself, not inferred from the commits:

| on 0.8.0 | |
|---|---|
| `cjose_jwe_decrypt(jwe, jwk, &len, NULL)` on any `ECDH-ES+A*KW` JWE | **SIGSEGV** — `err` is documented as optional |
| `{"kty":"RSA",...,"d":"AA"}` | imports as a **private** key; signs; the JWS does not verify; the export writes `"d":""`, which cjose cannot read back |
| `{"kty":"RSA",...,"d":"=="}` | imports as a public key |
| an RSA JWK carrying `"oth"` | silently used as if it had two primes |
| an `"epk"` carrying a private member | accepted, against RFC 7518 section 4.6.1.1 |
| a JSON JWE repeating a header name | accepted, and the per-recipient copy wins — so an unprotected `"alg"` shadows the protected one |

The first one is the reason I am writing: it is a crash on a documented-optional argument, reachable by any application that passes `NULL` and receives an `ECDH-ES+A*KW` JWE.

### What the branch carries

Fixes only. No algorithm, no API, no ABI change, and the OpenSSL 1.0.1 floor and the autotools build stay.

| commit | |
|---|---|
| #191 | the NULL `cjose_err` crash |
| #189, #192 | RSA private member present but valueless |
| #193 | the same for a value whose octets are all zero — **still open here**, and the reason the chain is otherwise incomplete |
| #190 | multi-prime `"oth"` refused |
| two ports | `epk` must be public; the header locations must be disjoint |

The last one needed hand-porting because both fixes were squash-merged together with things this line must not take: X25519/X448, A\*GCMKW, and the `crit` refusal. Deliberately left behind: the `crit` refusal, since rejecting input 0.8.0 accepts is not patch-release behaviour, and every feature and build change.

### What it will cost a user

Stated plainly, because a maintenance line should not surprise anyone:

- A JWE that repeats a member name across two header locations no longer imports. That is the point for `"alg"`, but it also refuses harmless repetitions such as `"kid"`. I know of no producer that does this — every JSON vector in `test/`, and the python-jwcrypto ones, are disjoint — but it is a behaviour change, and the CHANGELOG says so.
- Supplying your own `"epk"` to an ECDH-ES encryption is refused. The key agreement overwrote it in any case, so nothing that worked stops working, but a caller doing it now gets an error instead of silence.

### Testing

121 checks pass, CMake and autotools, with `CJOSE_ENABLE_RSA1_5` off and on; ASan, UBSan and valgrind clean; `clang-format` clean. Each of the six issues above is covered by a test that fails on `0.8.0`. Version bumped to 0.8.1 in both build systems with a CHANGELOG entry.

Two caveats I would rather state than have you find. No OpenSSL 1.x is installed on the machine I tested on, so the 1.0.2 and 1.1.1 paths are argued rather than run: the diff adds no OpenSSL identifier at all, only Jansson calls, and the set of OpenSSL symbols in `src/` is identical to 0.8.0's. And an independent review of this branch caught a real defect in my first cut of the hand-port — the encryption could assemble a JWE its own import refused — which is fixed in the last commit; I mention it because it is the kind of thing a hand-port gets wrong, and it is worth your eye on that commit in particular.

One unrelated thing the same review turned up: out-of-tree autotools builds cannot compile the test binary on 0.8.0, because `test/Makefile.am` passes only `-I$(top_builddir)/include`. One line, fixed on the branch, and it applies to master's ancestors too.

zandbelt:
Closes #194, now that `0.8.x` exists to target. Nine commits off the `0.8.0` tag, which is where this branch points, so it applies as a fast-forward.

Fixes only. No algorithm, no API, no ABI change; the OpenSSL 1.0.1 floor, the autotools build and `libcjose.so.0` all stay.

| commit | |
|---|---|
| #191 | `cjose_jwe_decrypt` dereferenced a NULL `cjose_err` on any `ECDH-ES+A*KW` JWE. A crash on a documented-optional argument, reachable by any application that passes `NULL` |
| #189, #192 | an RSA JWK whose private member is present but valueless imported as a key it could not then export |
| #193 | the same for a value whose octets are all zero. **Still open on `master`** — if it changes in review I will match this commit to it |
| #190 | a multi-prime RSA JWK (`oth`) was silently used as if it had two primes |
| two ports | an `epk` carrying a private member is refused (RFC 7518 section 4.6.1.1); the three JWE header locations must be disjoint (RFC 7516 section 7.2.1) |
| new | a caller-supplied `epk` is refused in the protected header too. The guard looked the parameter up as a string and an `epk` is an object, so it never saw one there |
| new | an **EC** JWK whose `d` is present but valueless is refused, the same defect #189 fixed for RSA and left for EC |

The last two needed hand-porting rather than a cherry-pick: upstream squash-merged both together with things this line must not take — X25519/X448, `A*GCMKW`, and the `crit` refusal.

**Deliberately left behind:** the `crit` refusal, since rejecting input that 0.8.0 accepts is not patch-release behaviour, and every feature and build change.

**What it costs a user**, since a maintenance line should not surprise anyone. All of these are in the CHANGELOG:

- A JWE repeating a member name across two header locations no longer imports. That is the point for `alg`, where the unauthenticated per-recipient copy won, but it also refuses harmless repetitions such as `kid`. Every JSON vector in `test/` and the python-jwcrypto ones are disjoint, so I know of no producer that does this.
- Supplying your own `epk` to an ECDH-ES encryption is refused. The key agreement overwrote it anyway, so nothing that worked stops working; a caller doing it now gets an error instead of silence.
- Three smaller ones follow from validating the shared unprotected header at encryption time, which 0.8.0 did not do — it validated the per-recipient header twice instead. `cjose_jwe_encrypt_multi` now refuses a `crit` that header carries and cannot process, which import already refused on 0.8.0; it now **accepts** an `alg` supplied only there, where 0.8.0 required it in the protected or per-recipient header; and an EC JWK with a present-but-valueless `d` no longer imports.

I would rather state that second one plainly than have it found in review: it is an acceptance-widening in a fixes-only release. It falls out of passing the right header to the disjointness check, and matches what merged upstream. Say the word and I will restrict it.

No version is bumped anywhere: `CMakeLists.txt`, `configure.ac` and the generated `configure` all still say 0.8.0, and the CHANGELOG section is headed `Unreleased` for you to number at release time.

## Testing

Each of the nine commits builds and passes on its own. On the tip: the full `check_cjose` run with `CJOSE_ENABLE_RSA1_5` off (122 checks) and on (121), no `clang-format` diff, valgrind clean, and every fix carries a regression test verified to fail on the `0.8.0` tag — including the headline crash, which I confirmed SIGSEGVs on a build of the tag itself.

`SONAME` is `libcjose.so.0` and the exported symbol list is byte-identical to 0.8.0. The autotools build works in tree and out of tree; `test/Makefile.in` is tracked on this line, so the out-of-tree include fix is applied there alongside `test/Makefile.am`.


zandbelt:
Pushed three corrections after running an independent review over the branch. Two of them were things I had claimed in the description above that were not true, so they are worth calling out rather than leaving in the diff:

- **The autotools build still said 0.8.0, and the out-of-tree fix did not work.** `configure` and `test/Makefile.in` are tracked generated files on this line. I bumped `configure.ac` and fixed `test/Makefile.am` without updating them, so `./configure` reported 0.8.0, `cjose.pc` and `version.h` would have installed as 0.8.0, and the out-of-tree test build still failed with exactly the error the commit says it fixes. Both are updated now, by hand rather than by `autoreconf`, since my autotools are 2.72/1.18.1 and these were generated with 2.69/1.16.1 — regenerating would have buried the change in thousands of unrelated lines. In-tree and out-of-tree `make check` both pass.

- **"Supplying your own `epk` is refused" was only two thirds true.** The guard looks the parameter up with `_cjose_jwe_get_from_headers`, which ends in `json_string_value` and returns NULL for anything that is not a string. An `epk` is an object, so it was never found and a caller's `epk` in the protected header was silently replaced instead of refused. The shared and per-recipient cases were caught after the fact by the disjointness check, which is why the tests passed. Fixed by backporting the JSON-valued lookup, with a test that fails without it.

- **The valueless-private-member fix was applied to RSA and left for EC.** `{"kty":"EC",...,"d":""}` and `"d":null` still imported as public keys on this branch — the identical #189 defect. The commit deferred it to #186, which has since merged upstream, so the deferral no longer holds. Fixed the same way the RSA import was.

I have also added three behaviour changes to the description and the CHANGELOG that I had not disclosed. One of them widens what cjose accepts rather than narrowing it, which deserves your attention in a fixes-only release: `cjose_jwe_encrypt_multi` will now produce a JWE whose `alg` sits only in the shared unprotected header, where 0.8.0 refused to. It falls out of passing the shared header to the disjointness check instead of the per-recipient one, which the check needs to work at all, and it matches what merged upstream — but if you would rather 0.8.x kept 0.8.0's stricter behaviour there, say so and I will add the restriction.

Nine commits now. Each builds and passes standalone; `SONAME` and the exported symbol list are unchanged from 0.8.0.

TheStormN (approved):

TheStormN (changes requested):
The PR review is good, but please remove all the versions bumps from both CMake and autotools as I'm doing that as part of the release process.

zandbelt:
Done, in f18ae38. `CMakeLists.txt`, `configure.ac` and the generated `configure` all say 0.8.0 again — `configure` is now byte-identical to the tag, so the whole diff is fixes and tests.

Two related things I did along with it, both easy to undo if you would rather:

- The CHANGELOG section is headed `Unreleased` rather than `0.8.1`, on the same reasoning: the number is yours to assign at release. The content is still there, because it is where the behaviour changes are written down and I pointed at it in the description above. Say the word and I will drop the section entirely and leave the notes to you.
- `test/Makefile.in` keeps its one change. That is not a version bump: it is the out-of-tree include path fix, and the file is tracked on this line, so `test/Makefile.am` alone would have had no effect. In-tree and out-of-tree `make check` both pass.

I also dropped "and release as 0.8.1" from that commit's subject, since it no longer does that.

Rebased rather than reverted, so no version bump is left anywhere in the history. Nine commits, each still building and passing on its own, 122 checks with `CJOSE_ENABLE_RSA1_5` off and 121 on.

TheStormN (approved):

PR #199, opened 2026-09-15
Record the 1.0.x changes and the 0.8.1 release in the CHANGELOG

zandbelt:
## Summary

Follow-up to your comment on #198. Documentation only, no code.

### The README already covers ML-DSA

#198 updated it: the JWS `alg` table has the `ML-DSA-44`, `ML-DSA-65`, `ML-DSA-87` row with `build option CJOSE_ENABLE_ML_DSA, OpenSSL >= 3.5`, the JWK `kty` table has the `AKP` row with the same requirement, the build options table has `CJOSE_ENABLE_ML_DSA`, and a paragraph says the algorithms are pure ML-DSA with the empty context string and that HashML-DSA is not offered.

The one place that did not mention it is **Prerequisites → Libraries**, which said `OpenSSL >= 3.0.0` and nothing more. That line now names the 3.5 requirement and points at the build option. It is the only README change here.

### The CHANGELOG had a larger gap than ML-DSA

The `1.0.0 (unreleased)` section held only the `Breaking` list that #187 wrote. Nothing merged after it had an entry — not ML-DSA, and not the nine PRs before it:

| | |
|---|---|
| #185 | PBES2-HS256+A128KW, PBES2-HS384+A192KW, PBES2-HS512+A256KW |
| #186 | X25519 and X448 ECDH-ES, the `epk` private-member refusal, the EC `d` refusal |
| #188 | A128GCMKW, A192GCMKW, A256GCMKW; the `crit` refusal; JWE header disjointness |
| #189, #190, #192, #193 | the RSA JWK import refusals |
| #191 | the NULL `cjose_err` crash in ECDH-ES+A\*KW decryption |
| #196 | `EVP_Q_mac` for HMAC |
| #198 | ML-DSA-44/65/87 and the AKP key type |

All of them are now recorded under `Update` and `Fix`. The entries reference pull requests, which is what the `Breaking` list of that section already does, rather than the commit links the released sections use. Nothing in the released sections is touched — the diff is 51 added lines and no deletions.

**Classification.** The `crit` refusal is under `Breaking`: it refuses a JWE or JWS that 0.8.0 accepted. Three more rules refuse input or calls that 0.8.0 accepted without changing the API, so they stay under `Fix` and a short `Compatibility` paragraph names them — the disjointness of the header locations, the refusal of a header parameter the algorithm generates, and the refusal of a valueless private member. That mirrors how the 0.8.1 notes handled the same rules.

### 0.8.1 was missing from this file entirely

You released 0.8.1 from the `0.8.x` branch on 2026-09-14, and its section only ever existed on that branch. On `main` the file jumped from the unreleased 1.0.0 straight to 0.8.0, so a released version was invisible in the release notes. Its section is copied over unchanged, including the autotools test-build fix, which describes what 0.8.1 shipped rather than what `main` has.

Comparing the two also turned up an entry the 1.0.0 list would otherwise have missed: the EC `d` refusal, which 0.8.1 lists separately and which `main` got in #186 rather than with the RSA members in #189.

## Testing

No code changes, so nothing to build or run. Verified instead that every entry is true of the tree at 8d1afeb: each function, macro, build option, header parameter and error code exists with that spelling, each behavioural claim matches the end state rather than an intermediate commit of a multi-commit PR (#188 scoped `crit` per algorithm family before refusing it outright; #185 changed how `p2s` is handled), and each PR attribution matches the commit that introduced the behaviour.


TheStormN (approved):

TheStormN:
Thanks!

